PR #22, opened 2017-07-13
Support parsing categorical (dictionary-encoded) columns from ipc arrow format

### LINKED ISSUE #7: Categorical column support
Need to support importing and creating categorical columns.

sklam:
Closes #7

This adds dependency on libarrow (`conda install -c conda-forge=0.5.*`).
The IPC parser API is changed to reflect new convention in mapd.  The schema and the payload is now in separate memory buffers.  The schema is expected to be in a CPU memory.  The payload is in GPU memory.

In addition, the buildscript is changed to pin libgdf_cffi version to libgdf version.


src/ipc.cu
```diff
@@ -157,33 +243,35 @@ protected:
...
-        auto header = parse_msg_header(header_buf);
-        if ( header.body_length > 0) {
-            throw ParseError("schema should not have a body");
-        }
-        parse_schema(header);
+        // Use Arrow to load the schema
+        const auto payload = std::make_shared<arrow::Buffer>(schema_buf, length);
+        auto buffer = std::make_shared<io::BufferReader>(payload);
+        std::shared_ptr<ipc::RecordBatchStreamReader> reader;
+        ipc::RecordBatchStreamReader::Open(buffer, &reader);
```
xhochy: You should check the return value of Open as this is used to indicate any errors.

wesm:
From the prior discussion in https://github.com/gpuopenanalytics/pygdf/issues/1, I am wondering it it would make packaging and testing much simpler to develop the Arrow GPU integration within the main Arrow codebase. Otherwise you may risk duplicating logic or missing other opportunities for code sharing (or generalizing the code that is already there in libarrow). Let me know how I can help -- I also have an NVIDIA graphics card and CUDA installed so I should be able to help with testing and debugging.

seibert:
I think we are likely to contribute some of this work back to Arrow once it is settled, but right now we're still iterating on how the IPC handshake should look with MapD, and generally experimenting with changes up and down the stack.    Once the approach is a little more figured out, we can definitely move some of this to Arrow.

wesm:
Makes sense. I've been following the commits both here and in MapD and it seems likely that there will soon be a need to share more code between MapD and pygdf, and one simple way would be as a leaf `libarrow_gpu` library. This also makes it easier for the rest of us to test Arrow CPU changes against your work to make sure that something doesn't get accidentally broken =) It would be great to see a community-standard Arrow-on-GPU solution for C and C++ users and to work together on maintaining shared build / packaging / CI. 

Note that there is no problem at all with experimental development inside the Arrow codebase so long as you don't break the build on master. I would rather see you all earning karma in the project so that you can be voted in sooner as committers. I'm not going to delay merging Arrow patches so long as the build isn't broken and there is some consensus amongst yourselves that the changes look good.

PR #25, opened 2017-07-25
Join

sklam:
* implements left, inner, outer join 
    * The output of these joins are indices of the left and right keys in the sequence of `L0, R0, L1, R1, L2, R2, ..., Ln-1, Rn-1, Ln, Rn`
    * Note: right join is not provided as it can be easily done as left join with swapped args in the higherlevel code.
* Adds moderngpu as a submodule


python/tests/test_joining.py
```diff
@@ -0,0 +1,229 @@
...
+
+from .utils import new_column, unwrap_devary, get_dtype
+
+
+@pytest.mark.parametrize('dtype', [np.int32, np.int64, np.float32, np.float64])
+def test_innerjoin(dtype):
+    cuda.current_context()
+    # Make data
+    left = np.array([0, 0, 1, 2, 3], dtype=dtype)
+    d_left = cuda.to_device(left)
```
seibert: There is a lot of duplicated code in the tests to make the left and right GDFs from the input arrays.  Could this be extracted into some helper functions to make the test cases more readable?

PR #26, opened 2017-07-26
Shuffle join result

sklam:
to return left and right indices in their own contiguous sequences.
This is more efficient for consumption.

### DIFF
--- a/src/joining.cu
+++ b/src/joining.cu
@@ -47,10 +47,6 @@ namespace {
 
 using namespace mgpu;
 
-std::ostream& operator << (std::ostream& out, int2 vec) {
-    return out << "(" << vec.x << ", " << vec.y << ")";
-}
-
 template <typename T>
 void dump_mem(const char name[], const mem_t<T> & mem) {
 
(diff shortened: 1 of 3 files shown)


src/joining.cu
```diff
@@ -122,27 +118,30 @@ mem_t<int> scan_join_bounds(const _join_bounds &bounds, int a_count, int b_count
...
+    int output_npairs = join_count + append_count;
+    mem_t<int> output(2 * output_npairs, context);
+    int* output_data = output.data();
 
     if (isInner){
         // Use load-balancing search on the segments. The output is a pair with
         // a_index = seg and b_index = lower_data[seg] + rank.
         auto k = [=]MGPU_DEVICE(int index, int seg, int rank, const int *lower) {
-            output_data[index] = make_int2(seg, lower[seg] + rank);
+            // output_data[index] = make_int2(seg, lower[seg] + rank);
```
seibert: Should this comment be removed?

src/joining.cu
```diff
@@ -157,7 +156,9 @@ mem_t<int2> compute_joined_indices(const _join_bounds &bounds,
             auto upper = get<1>(lower_upper);
             auto result = lower + rank;
             if ( lower == upper ) result = -1;
-            output_data[index] = make_int2(seg, result);
+            // output_data[index] = make_int2(seg, result);
```
seibert: Same here?

src/joining.cu
```diff
@@ -171,8 +172,11 @@ void outer_join_append_right(T *output_data,
                              const mem_t<int> &matches,
                              int append_count, int join_count,
                              context_t &context) {
+    int output_npairs = join_count + append_count;
     auto appender = [=]MGPU_DEVICE(int index, int seg, int rank) {
-        output_data[index + join_count] = make_int2(-1, seg);
+        // output_data[index + join_count] = make_int2(-1, seg);
```
seibert: And here?

sklam: good catch.  they should be removed

PR #45, opened 2018-06-14
[Review] Comparison operators

felipeblazing:
NOTE: stream compaction is still a work in progress
This pull request adds the following apis to functions.h

`gdf_error gpu_comparison_static_i8(gdf_column *lhs, int8_t value, gdf_column *output,gdf_comparison_operator operation);`

`gdf_error gpu_comparison_static_i16(gdf_column *lhs, int16_t value, gdf_column *output,gdf_comparison_operator operation);`

`gdf_error gpu_comparison_static_i32(gdf_column *lhs, int32_t value, gdf_column *output,gdf_comparison_operator operation);`

`gdf_error gpu_comparison_static_i64(gdf_column *lhs, int64_t value, gdf_column *output,gdf_comparison_operator operation);`

`gdf_error gpu_comparison_static_f32(gdf_column *lhs, float value, gdf_column *output,gdf_comparison_operator operation);`

`gdf_error gpu_comparison_static_f64(gdf_column *lhs, double value, gdf_column *output,gdf_comparison_operator operation);`

All of these functions are used to create a stencil. A stencil is basically an array of any type (we can choose later on to use a bitmask if we so choose) that has a 1 or a 0 in it to indiciate if a row in a column passed a filter or not,  So you pass in a gdf_column, a static value to compare to, and a comparison operation (you can see which ones are implemented in include/gdf/types.h). So the output is a gdf_column that has the stencil which results from processing this comparison against the column.

Note: the preferred output type here should be int8_t for compactness

`gdf_error gpu_comparison(gdf_column *lhs, gdf_column *rhs, gdf_column *output,gdf_comparison_operator operation);`

This api basically lets you do row wise comparison of two columns using the operation provided. It handles nulls and can take numeric types int8, int16, int32, float, double. You can mix and match inputs and output types at your own peril (like multiplying two int32s and storing the value in an int8_t)

`gdf_error gpu_apply_stencil(gdf_column *lhs, gdf_column * stencil, gdf_column * output);`

This function is used for stream compaction. The idea here is that you take a column and a stencil and return an output where all the values where stencil = true appear at the begining of the pointer inside of output. The size in output is set to the new size which depends on the stencil. Things that cause an element NOT to be copied over are having a 0 value in the stencil position or the valid bit for that row set to off.

NOTE: we are not doing anything with allocations here so it would be up to the person calling this api to ACTUALLY compact the allocations where output was stored


CMakeLists.txt
```diff
@@ -31,7 +46,18 @@ cuda_add_library(gdf SHARED
     src/joining.cu
     src/scan.cu
     src/segmented_sorting.cu
+    src/filterops.cu
+    src/bitmaskops.cu
+	src/streamcompactionops.cu
+    src/datetimeops.cu
 )
+if (NOT arrow)
```
sklam: With the setup using the conda environment, arrow is already included and available.  This check is not able to find the arrow build in conda.  Commenting out this block will "fix" it though.

include/gdf/cffi/functions.h
```diff
@@ -1,10 +1,14 @@
 /* column operations */
+#ifndef _GDF_FUNCTIONS_H_
```
sklam: Files under the `gdf/cffi` directory is to be parsed by python's `cffi` module, which doesn't support any preprocessor.  The headers in the directory is already included in `gdf/gdf.h`, thus it is not expected to be included directly.  So, the `#ifndef` is not needed.

include/gdf/cffi/functions.h
```diff
@@ -392,3 +412,29 @@ gdf_error gdf_max_f32(gdf_column *col, float *dev_result, gdf_size_type dev_resu
...
+//These compare every value on the left hand side to a static value and return a stencil in output which will have 1 when the comparison operation returns 1 and 0 otherwise
+gdf_error gpu_comparison_static_i8(gdf_column *lhs, int8_t value, gdf_column *output,gdf_comparison_operator operation);
+gdf_error gpu_comparison_static_i16(gdf_column *lhs, int16_t value, gdf_column *output,gdf_comparison_operator operation);
+gdf_error gpu_comparison_static_i32(gdf_column *lhs, int32_t value, gdf_column *output,gdf_comparison_operator operation);
+gdf_error gpu_comparison_static_i64(gdf_column *lhs, int64_t value, gdf_column *output,gdf_comparison_operator operation);
+gdf_error gpu_comparison_static_f32(gdf_column *lhs, float value, gdf_column *output,gdf_comparison_operator operation);
+gdf_error gpu_comparison_static_f64(gdf_column *lhs, double value, gdf_column *output,gdf_comparison_operator operation);
+
+//allows you two compare two columns against each other using a comparison operation, retunrs a stencil like functions above
+gdf_error gpu_comparison(gdf_column *lhs, gdf_column *rhs, gdf_column *output,gdf_comparison_operator operation);
```
sklam: Duplicated declaration.  `cffi` will complain.

include/gdf/cffi/types.h
```diff
@@ -1,3 +1,6 @@
+#ifndef _GDF_TYPES_H_
```
sklam: Same issue as above in `gdf/cffi/functions.h`

include/gdf/cffi/types.h
```diff
@@ -27,6 +32,7 @@ typedef struct gdf_column_{
     gdf_valid_type *valid;
     gdf_size_type size;
     gdf_dtype dtype;
+    gdf_size_type null_count = 0;
```
sklam: `cffi` can't handle the zero initializer.

src/bitmaskops.cu
```diff
@@ -0,0 +1,100 @@
+#include <gdf/gdf.h>
+#include <gdf/utils.h>
+#include <gdf/errorutils.h>
+#include <gdf/cffi/functions.h>
```
sklam: `gdf/cffi/functions.h` shouldn't be referenced directly and is already included by `gdf/gdf.h`

tests/filterops_numeric/helper/utils.cu
```diff
@@ -0,0 +1,192 @@
+
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

tests/filterops_numeric/helper/utils.cuh
```diff
@@ -0,0 +1,94 @@
+
+#ifndef GDF_TEST_UTILS
+#define GDF_TEST_UTILS
+
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

tests/filterops_numeric/test_example.cu
```diff
@@ -0,0 +1,80 @@
+#include "gtest/gtest.h"
+
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

tests/filterops_numeric/test_filterops.cu
```diff
@@ -0,0 +1,119 @@
+#include "gtest/gtest.h"
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

tests/filterops_numeric/test_filterops_augmented.cu
```diff
@@ -0,0 +1,145 @@
+#include "gtest/gtest.h"
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

tests/filterops_numeric/test_gpu_concat.cu
```diff
@@ -0,0 +1,53 @@
...
+ Copyright   : Your copyright notice
+ Description : Compute sum of reciprocals using STL on CPU and Thrust on GPU
+ ============================================================================
+ */
+
+#include "gtest/gtest.h"
+
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

tests/filterops_numeric/test_with_masks.cu
```diff
@@ -0,0 +1,130 @@
...
+ Description : Compute sum of reciprocals using STL on CPU and Thrust on GPU
+ ============================================================================
+ */
+
+#include "gtest/gtest.h"
+
+
+#include <iostream>
+#include <gdf/gdf.h>
+#include <gdf/cffi/functions.h>
```
sklam: see above

include/gdf/cffi/types.h
felipeblazing: ok we can drop that

include/gdf/cffi/types.h
aocsa: ok we will drop that

include/gdf/cffi/functions.h
aocsa: ok we will drop that too

.travis.yml
```diff
@@ -32,6 +32,7 @@ install:
   - conda --version
   # install conda build
   - conda install conda-build anaconda-client --yes
+  - conda install anaconda libboost --yes
```
scopatz: Shouldn't this be combined with the line above?

CMakeLists.txt
```diff
@@ -31,16 +90,21 @@ cuda_add_library(gdf SHARED
     src/joining.cu
     src/scan.cu
     src/segmented_sorting.cu
+    src/filterops.cu
+    src/bitmaskops.cu
+	src/streamcompactionops.cu
```
scopatz: Should have consistent indentation.

CMakeLists.txt
```diff
@@ -31,16 +90,21 @@ cuda_add_library(gdf SHARED
...
+    src/datetimeops.cu
+    src/sqls_ops.cu
+	src/hashops.cu
+	#src/windowedops.cu
```
scopatz: Should have consistent indentation.

cmake/Templates/Arrow.CMakeLists.txt.cmake
```diff
@@ -0,0 +1,66 @@
+#=============================================================================
```
scopatz: Why is this file needed? I am a little confused.

GPUtester:

mike-wendt:
ok to test

sklam:
My commit fixed the conda upload error.  But there is still a problem in the TravisCI during the download of `cuda-repo-ubuntu1404_8.0.61-1_amd64.deb`.

PR #47, opened 2018-06-18
[Review] Datetime operators

williamBlazing:
Have working extract year, month, day, hour, minute and second functions for date64, date32, timestamp (all resolutions s, ms, us, ns)

Have unit tests.


nsakharnykh:
Also I believe google tests should be updated according to https://github.com/gpuopenanalytics/libgdf/pull/60

PR #49, opened 2018-06-22
[Review] Multi-column group-by and filters - sort-based single-aggregation

aschaffer:
Multi-column single aggregation API redesign. Google Tests (GTest) in place now for all GROUP-BYs.


include/gdf/cffi/functions.h
```diff
@@ -392,3 +392,90 @@ gdf_error gdf_max_f32(gdf_column *col, float *dev_result, gdf_size_type dev_resu
...
+		     void** d_cols,    //out: pre-allocated device-side array to be filled with gdf_column::data for each column; slicing of gdf_column array (host)
+		     int* d_types,     //out: pre-allocated device-side array to be filled with gdf_colum::dtype for each column; slicing of gdf_column array (host)
+		     void** d_vals,    //in: device-side array of values to filter against (type-erased)
+		     size_t* d_indx,   //out: device-side array of row indices that remain after filtering
+		     size_t* new_sz);  //out: host-side # rows that remain after filtering
+
+gdf_error gdf_group_by_count(size_t nrows,     //in: # rows
+			     gdf_column* cols, //in: host-side array of gdf_columns
+			     size_t ncols,     //in: # cols
+			     int flag_sorted,  //in: flag specififying if rows are pre-sorted (1) or not (0)
```
mtjrider: Suggest adding an `enum` for `SORTED` or `UNSORTED` for use here. Would likely be useful to keep track of state variables that can be used across GDF operations and place these flags there.

mtjrider:
Apart from a cosmetic change to facilitate ease of use with respect to expanding stateful operation via `enum` or similar for various flags (e.g.) `SORTED`, this looks good to me.

Passes checks and builds without issue.

mtjrider:
Adding some additional commentary:

There is some concern regarding the proposed API for group-by. Some of the inputs/outputs aren't necessary for hash-based methods.

Could you provide a sample usage? Why do we need these outputs?

How can we generalized this API for multi-column multi-aggregators without efficiency?

aschaffer:
Usage example:
    template<typename T> using Vector = thrust::device_vector<T>;
    size_t ncols = 3;
    size_t& nrows = sz;
    
    Vector<void*> d_cols(ncols, nullptr);
    Vector<int>   d_types(ncols, 0);
    
    std::vector<gdf_column> v_gdf_cols(ncols);
    v_gdf_cols[0].data = static_cast<void*>(dc1.data().get());
    v_gdf_cols[0].size = sizeof(int);
    v_gdf_cols[0].dtype = GDF_INT32;

    v_gdf_cols[1].data = static_cast<void*>(di1.data().get());
    v_gdf_cols[1].size = sizeof(int);
    v_gdf_cols[1].dtype = GDF_INT32;

    v_gdf_cols[2].data = static_cast<void*>(dd1.data().get());
    v_gdf_cols[2].size = sizeof(double);
    v_gdf_cols[2].dtype = GDF_FLOAT64;

    gdf_column* h_columns = &v_gdf_cols[0];
    void** d_col_data = d_cols.data().get();
    int* d_col_types = d_types.data().get();

    int* ptr_d_indx = d_indx.data().get();
    int* ptr_d_kout = d_keys.data().get();
    int* ptr_d_vals = d_vals.data().get();

    size_t n_group = 0;
    int flag_sorted = 0;
    gdf_group_by_count(nrows, h_columns, ncols, flag_sorted, d_col_data, d_col_types, ptr_d_indx, ptr_d_kout, ptr_d_vals, &n_group);

gdf_group_by...() functions are just C-style wrappers (for Python binding reasons) around corresponding calls of functions from include/sqls_rtti_comp.hpp. Call them real-workers. The justification for the arguments is explained in that header. 

But, essentially, the array of gdf_columns is a host array (with elements that aggregate device pointers, among other things). In order to manipulate the set of columns on device, two device arrays must be pre-allocated and passed to the gdf_group_by...(): a void** d_columns, and a int* d_types. The caller is responsible for allocating them.

They need to be allocated by the caller so that allocation inside the real-workers can be avoided. And the reason to avoid that is because the real-workers use streams (that for the time being default to NULL stream). And allocation inside a stream would beat the purpose of the stream.

There's a C++ tester. It wasn't check in because a Python testers are desired, which are still work-in-progress. But I can look into cleaning up the C++ tester I have and checking it (them) in.

nsakharnykh:
I suggest to rename the pull request to something like "Multi-column filters and sort-based group-by"

include/gdf/cffi/functions.h
```diff
@@ -392,3 +392,90 @@ gdf_error gdf_max_f32(gdf_column *col, float *dev_result, gdf_size_type dev_resu
...
+   GROUP-BY
+ */
+gdf_error gdf_order_by(size_t nrows,     //in: # rows
+		       gdf_column* cols, //in: host-side array of gdf_columns
+		       size_t ncols,     //in: # cols
+		       void** d_cols,    //out: pre-allocated device-side array to be filled with gdf_column::data for each column; slicing of gdf_column array (host)
+		       int* d_types,     //out: pre-allocated device-side array to be filled with gdf_colum::dtype for each column; slicing of gdf_column array (host)
+		       size_t* d_indx);  //out: device-side array of re-rdered row indices
+
+gdf_error gdf_filter(size_t nrows,     //in: # rows
```
sklam: It's not obvious to me.  Is this function doing a "where equals"?  If so, a more descriptive name will be better.

include/gdf/cffi/functions.h
```diff
@@ -392,3 +392,90 @@ gdf_error gdf_max_f32(gdf_column *col, float *dev_result, gdf_size_type dev_resu
...
+/* 
+ Multi-Column SQL ops:
+   WHERE (Filtering)
+   ORDER-BY
+   GROUP-BY
+ */
+gdf_error gdf_order_by(size_t nrows,     //in: # rows
+		       gdf_column* cols, //in: host-side array of gdf_columns
+		       size_t ncols,     //in: # cols
+		       void** d_cols,    //out: pre-allocated device-side array to be filled with gdf_column::data for each column; slicing of gdf_column array (host)
```
sklam: I get that `d_cols` is a device allocation holding `ncols` of `void*`, but  I do not understand what "slicing of gdf_column array (host)" is referring to.

include/gdf/cffi/functions.h
```diff
@@ -392,3 +392,90 @@ gdf_error gdf_max_f32(gdf_column *col, float *dev_result, gdf_size_type dev_resu
...
+
+gdf_error gdf_group_by_sum(size_t nrows,     //in: # rows
+			   gdf_column* cols, //in: host-side array of gdf_columns
+			   size_t ncols,     //in: # cols
+			   int flag_sorted,  //in: flag specififying if rows are pre-sorted (1) or not (0)
+			   gdf_column agg_in,//in: column to aggregate
+			   void** d_cols,    //out: pre-allocated device-side array to be filled with gdf_column::data for each column; slicing of gdf_column array (host)
+			   int* d_types,     //out: pre-allocated device-side array to be filled with gdf_colum::dtype for each column; slicing of gdf_column array (host)
+			   int* d_indx,      //out: device-side array of row indices after sorting
+			   gdf_column agg_p, //out: reordering of d_agg after sorting; requires shallow (trivial) copy-construction (see static_assert below);
```
sklam: Will this contain `agg_in` column values by the groupby reordering?

williamBlazing:
approved

PR #52, opened 2018-06-26
Multi-column hash-join feature

nsakharnykh:
This pull request adds multi-column hash-based joins: core GPU implementation, low-level and high-level APIs, and functional tests. 

For single pair-wise joins it is possible to switch between sort and hash-based joins at compile time by using the `HASH_JOIN` cmake flag. For multi-column joins we're only using hash-based implementation (`HASH_JOIN` option must be ON). 

The CUDA kernels and the concurrent map class are located in `src/hash-join`, as well as the low-level API in `src/hash-join/join_compute_api.h`. The higher level APIs and memory management for hash-joins are in `src/joining.h`. The sort-based join implementation is moved to `src/sort-join.cuh`. The corresponding gdf_ functions that are used for cffi interfaces are located in `src/joining.cu`. This PR also introduces new error types for joins.

_Known limitations_: the current implementation of multi-column joins does not support more than 3 columns.


src/hash-join/managed.cuh
```diff
@@ -0,0 +1,19 @@
+#ifndef MANAGED_CUH
```
kkraus14: A copyright header should be added above this line.

src/hash-join/concurrent_unordered_multimap.cuh
```diff
@@ -0,0 +1,515 @@
...
+#include <type_traits>
+#include <cassert>
+
+#include <thrust/pair.h>
+
+#include "managed_allocator.cuh"
+#include "managed.cuh"
+#include "hash_functions.cuh"
+
+// TODO: replace this with CUDA_TRY and propagate the error
```
sklam: This TODO item is much needed.  Currently, CUDA_RT_CALL only prints the error message and keeps going, even though the error is likely unrecoverable.

src/joining.h
```diff
@@ -0,0 +1,133 @@
...
+
+  // output buffer
+  joined_type* joined = NULL;
+
+  cudaError_t error;
+  bool cont = true;
+  while (cont) {
+    // allocate an output buffer to store pairs, prefetch the estimated output size
+    CUDA_RT_CALL( cudaMallocManaged(&joined, sizeof(joined_type) * joined_size) );
+    CUDA_RT_CALL( cudaMemPrefetchAsync(joined, sizeof(joined_type) * joined_size, 0) ); // FIXME: use GPU device id from the context?
```
sklam: Any chance this FIXME is applied before this gets merged.

sklam:
It looks like I can remove all the `cudaMemPrefetchAsync` calls and the tests still passes.  Now, I can test on MBP.

src/hash-join/managed.cuh
nsakharnykh: @kkraus14 it does have the copyright, just not showing in the diff here for some reason

src/joining.h
nsakharnykh: Fixed with bcfcb27

src/hash-join/concurrent_unordered_multimap.cuh
nsakharnykh: I added an exit after we hit an error in bcfcb27

src/hash-join/managed.cuh
kkraus14: Resolved

PR #53, opened 2018-06-27
[Review] Added casting support for datetime types 

williamBlazing:
You can now cast any type to date64, date32 or timestamp and viceversa

Note that type casts between differing primitive data types behave like your typical type cast. Type casts between primitive type casts and datetime types behave like type casts between the actual primitive type and the primitive type underlying the datetime types (date64 and timestamp are int64, date32 is int32). Type casts between datetime types are logical type casts. For example date32 to timestamp(us) will have a completely different numberical value, but logically will represent the same day.

Created unit tests to test all casting combinations


scopatz:
CI seems to [be failing](https://travis-ci.org/gpuopenanalytics/libgdf/jobs/404990210#L5457) with 
`cffi.error.CDefError: cannot parse "gdf_dtype_extra_info dtype_info;"` So it seems that some of the new `dtype_info` options haven't been exposed to CFFI.

williamBlazing:
Fixed several issues that were causing CFFI to fail. Had to change the API a bit. I notice that we need to change  the location of the unit tests so that they reside where the foo-tests place holder is

scopatz:
There are now conflicts with master, sadly

PR #59, opened 2018-07-12
[Review] Multi-column hashing

nsakharnykh:
This PR adds the following features:
* a new function that hashes multiple columns together and creates a new column with the combined hash values
* minimal google tests infrastructure, google tests (`src/tests/hashing/hash_test.cu`) and python tests (`test_hashing.py`) for the new hashing API
* enables the hash-join implementation by default for all joins

Here is the proposed API for the multi-column hashing:

`gdf_error gdf_hash(int num_cols, gdf_column **input, gdf_hash_func hash, gdf_column *output)`


python/tests/test_hashing.py
```diff
@@ -0,0 +1,113 @@
...
+    
+    # hash_input6 = np.array(np.random.randint(0, 28, nrows), dtype=np.float64)
+    # hash_input6[-1] = hash_input6[0]
+    # hash_input.append(hash_input6)
+
+    # pytest on all test files fails if columns>5, but works well if called seperately like
+    # pytest --cache-clear -vvs ~/libgdf/python/tests/test_hashing.py
+
+    ncols = len(hash_input)
+    magic = 0
```
sklam: you can use `libgdf.GDF_HASH_MURMUR3` instead.

sklam:
The python C-API is good.

mtjrider:
Currently, there are merge conflicts. Please address these conflicts before.

What more work needs to be done to address the completion of this PR?

cmake/Templates/GoogleTest.CMakeLists.txt.cmake
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 2.8.12)
```
scopatz: We should probably be using at least CMake v3, at this point.

python/tests/test_hashing.py
```diff
@@ -0,0 +1,113 @@
...
+    
+    with _make_hash_input(hash_input, ncols) as (col_input):
```
scopatz: It is a little odd to have `col_input` in parens.

src/joining.h
```diff
@@ -17,6 +17,7 @@
 /* Header-only join C++ API (high-level) */
 
 #include <limits>
+#include <iostream>
```
scopatz: It seems as though we'd want as many of the includes as possible in the `*.cu` file, rather than in the header. It didn't look like anything else in joining.h was actually using iostream

nsakharnykh:
I rebased on the current master and made a few decorative changes

src/joining.h
nsakharnykh: fixed with 11a7eda

PR #60, opened 2018-07-13
[Review] CMake Improvements

aucahuasi:
- Build with g++-5.4
- Can run pytest (all tests pass)
- Update Readme.md documentation regarding the build process

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,30 +8,42 @@ The following instructions are tested on Linux and OSX systems.
 
 Compiler requirement:
 
-* gcc 4.8 (required to be ABI compatible with libarrow).
+* g++-4.8 or
+* g++-5.4
 
 ## Get dependencies
 
 **Note**: This repo uses submodules. Make sure you cloned recursively:
-```
+
+```bash
 git clone --recurse-submodules git@github.com:gpuopenanalytics/libgdf.git
 ```
+
 Or, after cloning:
-```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -8,30 +8,42 @@ The following instructions are tested on Linux and OSX systems.
...
 cd libgdf
 git submodule update --init --recursive
 ```
 
-It is recommended to setup a conda environment for the dependencies.
+Since cmake will download and build Apache Arrow you may need to install Boost C++:
+
+```bash
+$ sudo apt-get install libboost-all-dev
+```
```
scopatz: You should also be able to get the boost deps from conda-forge as well.

kkraus14: Additionally I don't think we can assume this is going to be running on a debian machine.

aucahuasi:
@jrhemstad Actually that is a good idea, I will create a symlink to libgdf/build/src/tests/

README.md
aucahuasi: Thanks for the feedback. I made a new PR https://github.com/gpuopenanalytics/libgdf/pull/64

aucahuasi:
PD: This new PR will make cmake can generate the google tests binaries inside gtests folder (nice idea!)

https://github.com/gpuopenanalytics/libgdf/pull/64

PR #61, opened 2018-07-17
Fix build. Ensure struct gdf_dtype_extra_info is C compatible

sklam:
PR #47 broke the build due to error with cffi.  This patch fixes it.


kkraus14:
Merging. Thanks @sklam!

PR #64, opened 2018-07-19
[Review] Make cmake generate c++ tests inside gtests folder

aucahuasi:
Also this PR improve the documentation based on the feedback of the PR https://github.com/gpuopenanalytics/libgdf/pull/60

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,8 +98,6 @@ if(HASH_JOIN)
   set(CUDA_NVCC_FLAGS  ${CUDA_NVCC_FLAGS};-DHASH_JOIN)
 endif()
 
-find_package(CUDA)
-
 # switch between cudaMalloc and cudaMallocManaged for hash tables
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -98,8 +98,6 @@ if(HASH_JOIN)
   set(CUDA_NVCC_FLAGS  ${CUDA_NVCC_FLAGS};-DHASH_JOIN)
 endif()
 
-find_package(CUDA)
-
```
scopatz: This doesn't seem correct to me.

aucahuasi: Is duplicated, see lines above

aucahuasi: That line was repeated without reason that is why i removed

scopatz: Oh yes, I see now. Sorry for the noise.

aucahuasi: No problem, thanks for taking the time to review this stuff ;)

PR #67, opened 2018-07-19
Draft CSV Reader

BradReesWork:


include/gdf/cffi/functions.h
```diff
@@ -481,3 +481,15 @@ gdf_error gdf_group_by_count(int ncols,                    // # columns
...
+
+/**  IO
+ *
+ *
+ */
+gdf_column ** read_csv(
+		const char 	*fileName, 			// in: the file to be loaded
+		const char 	delimiter,			// in: the delimiter
+		const int	num_cols,			// in: number of columns
+		const char 	**col_names,		// in: ordered array of column names
```
scopatz: The naming convention here seems inconsistent with `fileName` and `num_cols`

include/gdf/cffi/types.h
```diff
@@ -13,6 +16,8 @@ typedef enum {
     GDF_DATE32,   // int32_t days since the UNIX epoch
     GDF_DATE64,   // int64_t milliseconds since the UNIX epoch
     GDF_TIMESTAMP,// Exact timestamp encoded with int64 since UNIX epoch (Default unit millisecond)
+    GDF_CATEGORY,
+    GDF_STRING,
```
scopatz: Should these have comments like the others above.

src/io/csvReader.cu
```diff
@@ -0,0 +1,1230 @@
...
+    uint64_t *			field_bits;		// on-device: bitmap of where a field break is located (delimiter)
+    int	*				recPerChunck;	// on-device: Number of records per bitmap chunks
+    long * 				offsets;		// on-device: index of record - for multiple records per chunks it is starting index
+
+    char				delimiter;		// host: the delimiter
+    long				num_bytes;		// host: the number of bytes in the data
+    long				num_bits;		// host: the number of 64-bit bitmaps (different than valid)
+	long            	num_records;  	// host: number of records (per column)
+	int					num_cols;		// host: number of columns
+    vector<gdf_dtype>	dtypes;			// host: array of dtypes (since gdf_columns are not created until end)
```
scopatz: weird spacing above

src/io/csvReader.cu
```diff
@@ -0,0 +1,1230 @@
...
+	//---  load file -- using memory mapping
+	struct stat     st;
+	int				fd;
+	void * 			mapData;
+	size_t 			bytes;
+
+	fd = open(fileName, O_RDONLY );
+	if (fd < 0) {
+		std::cerr << "Error opening file" << std::endl;
+		//return GDF_C_ERROR;
```
scopatz: remove stale comment please.

scopatz: There are a few of these below, so I won't point them all out :)

src/io/csvReader.cu
```diff
@@ -0,0 +1,1230 @@
...
+
+
+/*
+ * What is passed in is the data type as a string, need to convert that into gdf_dtype enum
+ */
+gdf_dtype convertStringToDtype(std::string &dtype) {
+
+	gdf_dtype dt = GDF_invalid;
+
+	if (dtype.compare( "str") == 0) {
```
scopatz: non-blocking but It would be better if this could be reformulated as a switch-case, rather than a bunch of if-else blocks.  I believe that dtypes get assigned a unique integer that you can use.

PR #83, opened 2018-07-31
[REVIEW] extending gdf logical ops to datetime types

hhuuggoo:
I'm not sure if this was the correct way to do it but here goes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,12 @@ env:
 
 matrix:
   include:
-    - env: CUDA=8.0.61-1
     - env: CUDA=9.0.176_384.81
     - env: CUDA=9.1.85_387.26
-    - env: CUDA=8.0.61-1 BUILD_CFFI=1 PYTHON=3.6
-    - env: CUDA=8.0.61-1 BUILD_CFFI=1 PYTHON=3.5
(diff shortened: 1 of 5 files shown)


GPUtester:

scopatz:
This seems reasonable to me. Thanks for putting this in @hhuuggoo

hhuuggoo:
I'm having an issue with test_hashing locally - verifying if it's something I did or if my machine is setup weird

hhuuggoo:
I don't get it.

If I run `make pytest` then test_hashing fails for me, however if I run that test on it's own, with `pytest tests/test_hashing.py` then it passes

scopatz:
It sounds like you should open a separate issue for that....

hhuuggoo:
oooh

https://github.com/gpuopenanalytics/libgdf/blob/master/python/tests/test_hashing.py#L78

```
 # pytest for hashing fails if called collectively on folder and if number of columns>5, but works well if called seperately like
 # pytest --cache-clear -vvs ~/libgdf/python/tests/test_hashing.py

```

scopatz:
It looks like the hashing test is failing here, which is likely expected.

mike-wendt:
rerun tests

mt-jones:
This is failing because of 
https://github.com/gpuopenanalytics/libgdf/issues/77

Merging.

PR #87, opened 2018-08-02
[REVIEW] Fix warnings in sqls CUDA files

harrism:
Specifically: initializing dummy variables to zero; removing meaningless const type qualifiers; adding missing return statements; removing `__device__` specifier on function that passes `cout` to a thrust function. 

Note: warnings caused by ModernGPU using shfl rather than shfl_sync persist. This is being handled separately.


GPUtester:

kkraus14:
add to whitelist

PR #89, opened 2018-08-06
[REVIEW] Add Apache arrow 0.10 support

pearu:


GPUtester:

kkraus14:
add to whitelist

scopatz:
The code looks good to me @pearu! Thanks for putting this in.

Unfortunately, the gpuCI is failing to get the arrow version, it looks like:

```
15:03:11 -- Checking for one of the modules 'arrow'
15:03:11 CMake Error at /home/jenkins/.conda/envs/libgdf_dev/share/cmake-3.6/Modules/FindPkgConfig.cmake:646 (message):
15:03:11   None of the required 'arrow' found
15:03:11 Call Stack (most recent call first):
15:03:11   cmake/Modules/FindArrow.cmake:76 (pkg_search_module)
15:03:11   CMakeLists.txt:64 (find_package)
15:03:11 
15:03:11 
15:03:11 CMake Error at cmake/Modules/FindArrow.cmake:80 (math):
15:03:11   math cannot parse the expression: " * 10000 + * 100 + ": syntax error,
15:03:11   unexpected exp_TIMES, expecting exp_OPENPARENT or exp_NUMBER (2)
15:03:11 Call Stack (most recent call first):
15:03:11   CMakeLists.txt:64 (find_package)
15:03:11 
15:03:11 
15:03:11 -- ARROW_NUMERIC_VERSION=
```

pearu:
The gpuCI failure is most probably originated from missing pkg-config in gpuCI:
```
12:03:11 -- Could NOT find PkgConfig (missing:  PKG_CONFIG_EXECUTABLE) 
```
Could  `pkg-config` added to gpuCI package list?
If not, another way to determine arrow version needs to be implemented in `FindArrow.cmake`.

pearu:
FYI, all tests pass with recently released Apache Arrow 0.10.0.

README.md
```diff
@@ -49,6 +50,19 @@ or
 $ conda install -c omnia boost=1.58.0=py33_0
 ```
 
+Libgdf supports Apache Arrow versions 0.7.1 (default) and 0.8+ that
+use different metadata versions. So, it is important to specify which
+Apache arrow version will be used during building libgdf.  To select
+required Apache Arrow version, define the following environment
+variables (using Arrow version 0.9.0 as an example):
```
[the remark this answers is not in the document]
pearu: @scopatz Actually, you are right. Otherwise, CI instances are using arrow 0.7.1. It might be reasonable to hold merging until pygdf is arrow 0.10.0 ready.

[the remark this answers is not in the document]
scopatz: I don't think it is worth holding off, we can always update this value for new arrow releases.

scopatz:
That is excellent news about 0.10.0!

README.md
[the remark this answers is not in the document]
kkraus14: It looks like in other places in the PR we're defaulting to 0.10.0 which I believe was released ~2 days ago. I'm in favor of having arrow 0.10.0 the default.

README.md
```diff
@@ -49,6 +50,19 @@ or
...
+```bash
+$ export ARROW_VERSION=0.9.0
```
kkraus14: Bump to Arrow 0.10.0.

CMakeLists.txt
```diff
@@ -129,8 +129,8 @@ cuda_add_library(gdf SHARED
     #src/windowedops.cu
 )
 
-
-target_link_libraries(gdf arrow)
+# boost_regex may be necessary only when arrow==0.9.0
+target_link_libraries(gdf arrow boost_regex)
```
kkraus14: Is this different with 0.10.0?

pearu: Yes, boost_regex is not needed for arrow 0.10.0 (tested). This issue is related to https://issues.apache.org/jira/browse/ARROW-2903 , in summary, linking against boost_regex is required for arrow versions `>=0.8` and `<=0.9.0`.

Note that the change in
https://github.com/gpuopenanalytics/libgdf/pull/89/files#diff-281a0db11f8c4d7c230f268619adb8bbR57
has the same origin.

Modifying these parts depends on the question, do we want to support arrow versions from 0.7.1 upto to the latest (0.10 currently) or can we drop support for some intermediate versions (such as 0.8, 0.9)?

kkraus14: Can we make a conditional check for if Arrow is specified as one of those versions we add boost_regex, otherwise we don't? I think supporting 0.8 - 0.9.0 would be good, but we should avoid including boost_regex unnecessarily.

pearu: Yes, we can. I will do it.

kkraus14:
Looks good to me! Will merge when @scopatz approves.

CMakeLists.txt
```diff
@@ -129,8 +129,12 @@ cuda_add_library(gdf SHARED
     #src/windowedops.cu
 )
 
-
-target_link_libraries(gdf arrow)
+if (ARROW_NUMERIC_VERSION EQUAL 900)
```
scopatz: Should this be greater than  or equal to?

pearu: Actually, just equal to. Only arrow 0.9 had this issue.

pearu:
@kkraus14 please hold on merging.
I have one fix coming up that makes all pygdf tests pass with arrow 0.10.0.

CMakeLists.txt
scopatz: Ahh OK, great.

scopatz:
This LGTM, once the pygdf fix is in! Thanks a ton @pearu!

pearu:
@kkraus14 @scopatz please review.

Btw, pygdf is now also arrow 0.10 ready: https://github.com/gpuopenanalytics/pygdf/pull/200

kkraus14:
@pearu Can we remove the 0.7.1 Arrow hardcoded data from this PR as well and use the programmatically generated data?

pearu:
@kkraus14, arrow hardcoded data is removed from tests.
Created JIRA RAPIOP-79 for adding pyarrow dependency.

kkraus14:
@pearu Thanks. This looks good to me, the travis CI failures are from networking issues on Travis's end it looks like. @scopatz @mt-jones Would you review?

scopatz:
This LGTM!

Issue #72, opened 2018-07-23
CUDA Compute Capability improperly set


mt-jones:
From CMakeLists.txt:

```
IF(CUDA_VERSION_MAJOR GREATER 7)
    set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-gencode;arch=compute_60,code=sm_60)
ENDIF(CUDA_VERSION_MAJOR GREATER 7)

IF(CUDA_VERSION_MAJOR GREATER 8)
    set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-gencode;arch=compute_70,code=sm_70)
ENDIF(CUDA_VERSION_MAJOR GREATER 8)
```
Virtual GPU architecture, and generated code (SASS) are hard-coded. The project will compile properly, but if you're not using (essentially) Volta / V100, you'll get errors like this:

`C++ exception with description "parallel_for failed: no kernel image is available for execution on the device" thrown in the test body.`

or similar.

Virtual GPU architecture and generated code should be discovered by the project. Or the target platforms should be expanded beyond `sm_60`, `sm_70`

mike-wendt:
@mt-jones is this fixed with the #59 merge?

PR #92, opened 2018-08-12
[WIP] Automatic Detection of GPU Architecture Flags

mt-jones:
There are no material changes to any source code, features, or functionality. The only changes made were to address issue https://github.com/gpuopenanalytics/libgdf/issues/72

Changes to CMake were made to enable automatic detection of the system's GPU architecture with the ability to specify fully compatible targets (Pascal, Volta). There were some simple utilities added to facilitate the detection of architecture and compute capability.


GPUtester:

kkraus14:
add to whitelist

mt-jones:
TravisCI does not have a GPU, and will report failures every time.

Running `cmake ..` will compile a simple program that uses the CUDA Runtime API to poll for device properties. Those device properties are then used to define the compiler flags.

I've switched the header to **[REVIEW]**. I've tested this program on Maxwell, Pascal, Volta.

On architectures preceding Pascal, certain features are nonfunctional.

mt-jones:
We do have some options to get rid of the failed TravisCI.

I could make the default build use Volta architecture options, and the new detector functionality can be made available with a user provided switch (e.g.) `cmake -DGPU_ARCH=detected ..`

Let me know what how you'd like to proceed.

kkraus14:
> We do have some options to get rid of the failed TravisCI.
> 
> I could make the default build use Volta architecture options, and the new detector functionality can be made available with a user provided switch (e.g.) `cmake -DGPU_ARCH=detected ..`
> 
> Let me know what how you'd like to proceed.

I am in favor of having the build default to Volta if it can't detect the architecture. In the case of Travis we could set up a build matrix to test building for Volta, Pascal, Maxwell, and Kepler if we want.

scopatz:
The code looks good to me, but I think it makes sense to try to get travis to pass before merging, so that this doesn't mask real errors in future PRs.

mt-jones:
Canonical usage assuming you're in a `build` subdirectory of the project:

Default compiles for Volta architecture:
```
cmake ..
```

Compile for Pascal:
```
cmake -DGPU_ARCH=pascal ..
```

Compile for Pascal and Volta:
```
cmake -DGPU_ARCH=pascal,volta ..
```

Compile with custom flags:
```
cmake -DGPU_ARCH=custom -DCUDA_NVCC_FLAGS= --generate-code=arch=compute_XX,code=sm_XX;...;--expt-extended-lambda;-std=c++11 ..
```

Noting that `-DGPU_ARCH=XXXX` is case **insensitive**, and that specifying custom flags requires you to add the necessary `--expt-extended-lambda;-std=c++11` yourself.

kkraus14:
Would it be cleaner to try to detect the architecture and if it fails default to Volta instead of having it all user defined?

mt-jones:
Why would it fail to default to Volta?

kkraus14:
To clarify I'm imagining the workflow as:
1. Try to detect the GPU Architecture and build appropriately
2. If 1 fails then default to building Volta architecture or Volta + Pascal architectures (opinions would be welcome here)

harrism:
What will be the mininum supported arch? I think the Default should
include that plus PTX to make sure it JITs.
I worry about making the default to autodetect. Do we not expect it to
be common to build on a different machine than run? I'm thinking about
cluster head nodes without GPUd installed. If default is autodetect and
you build on a head node of a Pascal cluster with no GPU, then it
compiles for Volta, you are going to get user angst.
If we expect GPU-naive users to do the building, might want to make the
default very inclusive (e.g. all supported arch)?
Maybe I'm overly concerned about user experience.


On Mon, Aug 13, 2018, at 12:11 PM, Keith Kraus wrote:
> To clarify I'm imagining the workflow as:


>  1. Try to detect the GPU Architecture and build appropriately
>  2. If 1 fails then default to building Volta architecture or Volta +
>     Pascal architectures (opinions would be welcome here)> — You are receiving this because you commented. Reply to this email
> directly, view it on GitHub[1], or mute the thread[2].
Links:

  1. https://github.com/gpuopenanalytics/libgdf/pull/92#issuecomment-412630727

kkraus14:
@harrism this is the exact kind of opinions I was looking for, thanks! The default was going to be to detect and build for detected architecture, but if there's no GPU the fallback would be to build for all of the supported architectures.

It sounds like you think we should default to all the supported architectures which we'd need to have a build for to distribute via Conda / Pip / other platforms anyway. That works for me. @mt-jones any objections?

mt-jones:
So, is it sufficient to have `cmake ..` try to detect the architecture, and if it fails, compile for all? (e.g.) Pascal, Volta.

kkraus14:
It sounds like we're leaning towards the default regardless being to compile for all supported architectures to allow compatibility to distribute it to additional nodes in a cluster. I'd still like to hear other opinions as well.

CMakeLists.txt
```diff
@@ -56,20 +57,159 @@ set_package_properties(
...
     # set warnings as errors
-    set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS};-Werror cross-execution-space-call;-Xcompiler -Wall,-Werror)
+    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Werror cross-execution-space-call -Xcompiler -Wall,-Werror")
 
     message(STATUS "CUDA_NVCC_FLAGS: ${CUDA_NVCC_FLAGS}")
 else()
     message(FATAL_ERROR "CUDA not found, please check your settings.")
 endif()
 
+if (CUDA_FOUND)
```
harrism: This is a lot of code. Recommend moving it to a separate file included from CMakeLists.txt

mike-wendt:
Will be solved with refactoring by @mt-jones

PR #93, opened 2018-08-13
[REVIEW] README Updates for CMake Install and Formatting Cleanup [skip-ci]

mike-wendt:
Changes to README:

- Include instructions on how to install the latest CMake
- Moved `git pull` instructions to the top
- Labeled the dependencies section for easier readability
- Promoted the `Configure and Build` section to the next highest heading level

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,11 +112,12 @@ development includes/changes the depedencies.  To do so, run:
 ```bash
 conda env update --name libgdf_dev --file ../conda_environments/dev_py35.yml
 ```
+
 Note that `dev_py35.yml` uses the latest version of pyarrow.
 Reinstall pyarrow if needed using `conda install
 pyarrow=$ARROW_VERSION -c conda-forge`.
 
-### Configure and build
+## Configure and Build
 
 This project uses cmake for building the C/C++ library. To configure cmake,
 run:


README.md
```diff
@@ -21,18 +36,18 @@ You can obtain CUDA from [https://developer.nvidia.com/cuda-downloads](https://d
 
 ### Get dependencies
 
-**Note**: This repo uses submodules. Make sure you cloned recursively:
+#### CMake
 
+Install the latest version of `cmake` using `pip`:
 ```bash
-git clone --recurse-submodules git@github.com:gpuopenanalytics/libgdf.git
+pip install cmake
```
scopatz: cmake is also on conda-forge

kkraus14: I think we should give both options as there will be users installing libgdf and pygdf from source without using conda.

mike-wendt: That is true, and it is mentioned further down in the doc about using a conda environment that will install cmake and arrow for you. 

The issue with the conda process is:

1. it is at the end of the README and if that is the preferred path for build and test then it should be upfront first
2. the build environment that is defined in `conda_environments/dev_py35.yml` is outdated and has not been updated to the same standard as the information on the README or what is used in the build environments

Given those reasons I stuck with focusing on installing without conda, and welcome input on fixing the other problem of defining preferred means to build and test libgdf.

Additionally Ubuntu doesn't support any CMake version except v3.5 so pip is one way to get around the trouble of installing the latest version and being operating system independent.

kkraus14: > it is at the end of the README and if that is the preferred path for build and test then it should be upfront first

I think we should have conda as the recommended way of installing dependencies, followed by pip, followed by apt/yum/etc installs

> the build environment that is defined in conda_environments/dev_py35.yml is outdated and has not been updated to the same standard as the information on the README or what is used in the build environments

Yes, this needs to be addressed, and if it's addressed we can have at the top of the readme they can create a conda environment from the file and it automagically takes care of everything.

nsakharnykh:
Should we merge this one? @mike-wendt @kkraus14

kkraus14:
> Should we merge this one? @mike-wendt @kkraus14

The comment above was never addressed, which should be done before merging.

mike-wendt:
This is a dead branch and with the merged cuDF repo these docs will change anyway. Closing PR w/o merge

PR #94, opened 2018-08-13
[WIP] Binary Operators

wmalpica:
Implemented so far is the first implementation concept using nested templating. It has been implemented with support for literals and 14 different binary operators. 

We would like feedback on the API implemented, which is:

``
gdf_error gdf_scalar_operation(gdf_column* out, gdf_column* vax, gdf_scalar* vay, gdf_scalar* def, gdf_binary_operator ope);

gdf_error gdf_vector_operation(gdf_column* out, gdf_column* vax, gdf_column* vay, gdf_scalar* def, gdf_binary_operator ope);

enum gdf_binary_operator {
    GDF_ADD,
    GDF_SUB,
    GDF_MUL,
    GDF_DIV,
    GDF_TRUE_DIV,
    GDF_FLOOR_DIV,
    GDF_MOD,
    GDF_POW,
    //GDF_COMBINE,
    //GDF_COMBINE_FIRST,
    //GDF_ROUND,
    GDF_EQUAL,
    GDF_NOT_EQUAL,
    GDF_LESS,
    GDF_GREATER,
    GDF_LESS_EQUAL,
    GDF_GREATER_EQUAL,
    //GDF_PRODUCT,
    //GDF_DOT
};

struct gdf_scalar {
    union gdf_data {
        std::int8_t  si08;
        std::int16_t si16;
        std::int32_t si32;
        std::int64_t si64;
        float        fp32;
        double       fp64;
    };

    gdf_data  data;
    gdf_dtype type;
};
``

We have already started on a second implementation using NVRTC which uses the same API.

Unit tests have not yet been implemented and will be implemented after an implementation method has been decided.

This current implementation right now only supports the second operand being able to be a scalar, which can be improved upon easily later.

In order to compile the binary operation functionality, use the cmake variable "BINARY_OPERATION_VERSION" and choose the binary version.

Example:
cmake -DCMAKE_BUILD_TYPE=Release -DBINARY_OPERATION_VERSION:STRING=V1 ../../code/libgdf

This cmake switch was added right now because compilation time is very large due to all the functions generated by the templated code. For example compilation on a laptop with an 8-core i7 processor and 16Gb ram was:
    Time to compile: 77m10.996s
    Ram memory used: More than 13.5GB
    Release binary size: 69,931,016 bytes


scopatz:
This has a conflict with master now

src/binary/binary2/kernel.cpp
```diff
@@ -0,0 +1,141 @@
...
+    void shiftMask(uint32_t& mask) {
+        #pragma unroll
+        for (int offset = 16; offset > 0; offset /= 2) {
+            mask += __shfl_down_sync(WARP_MASK, mask, offset);
+        }
+    }
+
+    template <typename TypeOut, typename TypeVax, typename TypeVay, typename TypeOpe>
+    __global__
+    void kernel_v_s(int size, TypeOut* out_data, TypeVax* vax_data, gdf_data vay_data) {
```
harrism: Shouldn't `pdf_data` on this line be `TypeVay`?

include/gdf/cffi/types.h
```diff
@@ -16,6 +20,23 @@ typedef enum {
     N_GDF_TYPES, /* additional types should go BEFORE N_GDF_TYPES */
 } gdf_dtype;
 
+union gdf_data {
```
harrism: Are the names in this union required to only be four characters for some reason? It would be nice to spell them out better, especially "invd", "tmst", "dt32" and "dt64".

src/binary/binary2/kernel.cpp
```diff
@@ -0,0 +1,141 @@
...
+        int start = tid + blkid * blksz;
+        int step = blksz * gridsz;
+
+        for (int i=start; i<size; i+=step) {
+            AbstractOperation<TypeOpe> operation;
+            out_data[i] = operation.template operate<TypeOut, TypeVax, TypeVay>(vax_data[i], vay_data[i]);
+        }
+    }
+
+    template <typename TypeOut, typename TypeVax, typename TypeVay, typename TypeDef, typename TypeOpe>
```
harrism: Wow, `TypeDef` is very risky typename, considering its proximity to `typedef`. Could just use TDef, TVax, etc. Or reorder, putting `Type` last in the names.

src/binary/binary2/kernel.cpp
```diff
@@ -0,0 +1,141 @@
...
+    __global__
+    void kernel_v_s_d(int size, gdf_data def_data,
```
harrism: Should `gdf_data` on this line be `TypeDef`?

src/binary/binary2/kernel.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+            TypeVax vax_data_aux = vax_data[i];
+            if ((is_vax_valid & mask) != mask) {
+                vax_data_aux = (TypeDef)def_data;
+            }
+
+            AbstractOperation<TypeOpe> operation;
+            out_data[i] = operation.template operate<TypeOut, TypeVax, TypeVay>(vax_data_aux, (TypeVay)vay_data);
+
+            __syncwarp();
```
harrism: What is the reason for `__syncwarp()` here? I don't see any sharing of data between threads. Is it necessary? Same question for all the other `__syncwarp()` calls.

src/binary/binary2/launcher.cpp
```diff
@@ -0,0 +1,115 @@
...
+    Launcher& Launcher::instantiate(gdf_column* out, gdf_column* vax, gdf_scalar* vay, gdf_binary_operator ope) {
+        arguments.clear();
+        arguments.push_back(gdf::getTypeName(out->dtype));
+        arguments.push_back(gdf::getTypeName(vax->dtype));
+        arguments.push_back(gdf::getTypeName(vay->dtype));
+        arguments.push_back(gdf::getOperatorName(ope));
+        return *this;
+    }
+
+    Launcher& Launcher::instantiate(gdf_column* out, gdf_column* vax, gdf_column* vay, gdf_binary_operator ope) {
```
harrism: Since each of these `instantiate` functions varies only in the type of one argument, and their bodies are identical, couldn't you template them to reduce redundancy? You might even use variadic templates to collapse all into one instantiate and one launch function.

src/binary/binary2/operation.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    template <typename ConcreteOperation>
+    struct AbstractOperation {
+        template <typename TypeOut, typename TypeVax, typename TypeVay>
+        __device__
+        TypeOut operate(TypeVax x, TypeVay y) {
+            return static_cast<ConcreteOperation*>(this)->template operate<TypeOut, TypeVax, TypeVay>(x, y);
+        }
+    };
+/*
```
harrism: Do you need to keep the commented-out code?

src/binary/binary2/traits.cpp
```diff
@@ -0,0 +1,216 @@
+namespace gdf {
+namespace cuda {
+
+const char* traits =
+R"***(
+#pragma once
+
+    struct IntegralSigned {};
```
harrism: Hmmmm, I believe Jitify already has some support for std::type_traits, e.g. https://github.com/NVIDIA/jitify/blob/master/jitify.hpp#L1124
Do you need to roll your own?

src/tests/binary-operation/library/operation.h
```diff
@@ -0,0 +1,91 @@
...
+        }
+    };
+
+    template <typename TypeOut, typename TypeVax, typename TypeVay>
+    struct Pow {
+        TypeOut operator()(TypeVax vax, TypeVay vay) {
+            return (TypeOut)pow((double)vax, (double)vay);
+        }
+    };
+}
```
harrism: It's nice in this situation (closing brackets far from their openings) to comment what is being closed, e.g.:
```
} // namespace operation
} // namespace test
} // namespace gdf
```

harrism (changes requested):
My biggest concern with this PR is that there is a lot of code and basically zero documentation. First, I think it needs high-level documentation of the functionality and organization, and it needs operational documentation for the main APIs being introduced (both external and internal APIs).

src/binary/binary2/kernel.cpp
ironbit: The operands types are column and scalar. gdf_data (scalar - union) is passed to the kernel. Inside the kernel, it is used a type cast operator when the operation is performed.

include/gdf/cffi/types.h
ironbit: There isn't a particular reason. Is there any name style and code style for this project?

src/binary/binary2/operation.cpp
ironbit: I implemented this version as a "prototype". It requires code style, name style, code organization, better testing, etc. Whether a requirement could be underflow detection, the commented code would be useful.

src/binary/binary2/traits.cpp
ironbit: I had some issue with that. I expressed it in my daily scrum. It is supposed to be working, but it isn't. I included the libraries, however "common_type" and others aren't working, because of the error messages using nvrtc.

src/binary/binary2/kernel.cpp
ironbit: No problem, it will be changed.

src/tests/binary-operation/library/operation.h
ironbit: Documentation is important. I was focused on implementation.

src/binary/binary2/launcher.cpp
ironbit: That's a good point

src/binary/binary2/kernel.cpp
ironbit: It's not correct. It'll be removed.

include/gdf/cffi/types.h
harrism: Human readable variable names is good practice for every project.

src/binary/binary2/traits.cpp
harrism: Let's work with Kate Clark and Ben Barsdell on making Jitify better. Are you in touch with them?

ironbit: No. I'm not. I remember having issues related to "type_traits" header and with compiler option "--device-as-default-execution-space", which wasn't working.

harrism: I will contact them.

include/gdf/cffi/types.h
```diff
@@ -91,6 +116,29 @@ typedef enum {
...
+    GDF_SUB,
+    GDF_MUL,
+    GDF_DIV,
+    GDF_TRUE_DIV,
+    GDF_FLOOR_DIV,
+    GDF_MOD,
+    GDF_POW,
+    //GDF_COMBINE,
+    //GDF_COMBINE_FIRST,
+    //GDF_ROUND,
```
scopatz: I don't think that it makes sense to have intermediate parts of an enum commented out.

scopatz: This can lead to API instabilities.

ironbit: I'm so sorry, I hadn't seen your message.
It is used only as an enum (not casting to integers).
It could be error-prone. In any case, it will be moved to the end.

nsakharnykh:
what is the corresponding pygdf PR for the bindings update?

cmake/Templates/GoogleBenchmark.CMakeLists.txt.cmake
```diff
@@ -0,0 +1,36 @@
+#=============================================================================
```
nsakharnykh, line 1: Can we use this for CI? @mike-wendt

include/gdf/cffi/types.h
```diff
@@ -120,6 +159,36 @@ typedef enum {
...
+    GDF_FLOOR_DIV,
+    GDF_MOD,
+    GDF_POW,
+    GDF_EQUAL,
+    GDF_NOT_EQUAL,
+    GDF_LESS,
+    GDF_GREATER,
+    GDF_LESS_EQUAL,
+    GDF_GREATER_EQUAL,
+    //GDF_COMBINE,
```
nsakharnykh, line 225: Add a description/comment why those are not implemented

src/binary/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+
+message(STATUS "Binary Operation Version: V2 - NVRTC")
+
+set(source_files
+    src/binary/binary2/binary.cpp
```
nsakharnykh: Rename `binary2` folder to `binary_nvrtc` or `binary_jitify`

src/binary/binary2/kernel.cpp
```diff
@@ -0,0 +1,155 @@
...
+    void shiftMask(uint32_t& mask) {
+        #pragma unroll
+        for (int offset = 16; offset > 0; offset /= 2) {
+            mask += __shfl_down_sync(WARP_MASK, mask, offset);
+        }
+    }
+
+    template <typename TypeOut, typename TypeVax, typename TypeVay, typename TypeOpe>
+    __global__
+    void kernel_v_s(int size, TypeOut* out_data, TypeVax* vax_data, gdf_data vay_data) {
```
nsakharnykh: We need `kernel_s_v` as well - should be 6 combinations overall, one kernel per the top level GDF function

src/binary/binary2/launcher.cpp
```diff
@@ -0,0 +1,94 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "binary/binary2/launcher.h"
+#include "binary/binary2/cuda.h"
+
+namespace gdf {
+    static thread_local jitify::JitCache JitCache;
```
nsakharnykh: We should be able to store JitCache when exiting the process to store in a file, then use an env variable to read it from the specified file to have the cache enabled between different processes

src/binary/binary2/operation.cpp
```diff
@@ -0,0 +1,213 @@
...
+    struct Add : public AbstractOperation<Add> {
```
nsakharnykh: Should add a comment with the description why this could be useful, i.e. for the overflows - if we ever wanted to enable this "feature" in the future

src/library/field.h
```diff
@@ -0,0 +1,109 @@
+/*
```
nsakharnykh: Move `library` under `bench/binary` or `tests/binary` since it's only used in tests, probably also rename to something like binary_helper

src/bench/binary-operation-benchmark.cpp
```diff
@@ -0,0 +1,203 @@
+/*
```
nsakharnykh: Needs to be moved to `src/bench/binary` to have a folder per libgdf operator (one folder for binary, one for parquet, one for joins, etc.)

src/binary/binary2/kernel.cpp
```diff
@@ -0,0 +1,155 @@
...
+        int blkid = blockIdx.x;
+        int blksz = blockDim.x;
+        int gridsz = gridDim.x;
+
+        int start = tid + blkid * blksz;
+        int step = blksz * gridsz;
+
+        for (int i=start; i<size; i+=step) {
+            AbstractOperation<TypeOpe> operation;
+            out_data[i] = operation.template operate<TypeOut, TypeVax, TypeVay>(vax_data[i], (TypeVay)vay_data);
```
nsakharnykh: We need to set the output valid bit mask - I don't see it being handled in the code. It should be an OR between the two bit masks of the two input operands.

nsakharnykh (changes requested):
We did a live PR review with @jrhemstad @williamBlazing and @ironbit - requesting more changes. Also, we need the python bindings updated to use the new binary ops API: @mt-jones, can you post an update on where we are with this?

src/binary/CMakeLists.txt
ironbit: The source code has been reorganized, that's why I think it's convenient to reply to every observation.
https://github.com/BlazingDB/libgdf/tree/binary-operators-draft/src/binary-operation/jit

src/binary/binary2/operation.cpp
ironbit: The source code has been optimized for jit.
The commented code has been moved outside the jit code and it has been added a description.
https://github.com/BlazingDB/libgdf/blob/96012d91aff817e6105a8fa8983e1587d237493f/src/binary-operation/jit/code/operation.cpp#L159

include/gdf/cffi/types.h
scopatz: Or, preferably, please remove.

src/binary-operation/jit/util/type.cpp
```diff
@@ -0,0 +1,96 @@
...
+                return "TrueDiv";
+            case GDF_FLOOR_DIV:
+                return "FloorDiv";
+            case GDF_MOD:
+                return "Mod";
+            case GDF_POW:
+                return "Pow";
+            //case GDF_COMBINE:
+            //case GDF_COMBINE_FIRST:
+            //case GDF_ROUND:
```
scopatz, line 74: Are these simply not implemented? Probably should remove or make a comment

src/bench/binary-operation-benchmark.cpp
ironbit: https://github.com/BlazingDB/libgdf/tree/binary-operators-draft/src/bench/binary-operation

src/library/field.h
ironbit: https://github.com/BlazingDB/libgdf/tree/binary-operators-draft/src/tests/binary-operation/util

src/binary/binary2/kernel.cpp
ironbit: The output valid bitmask is processed in all kernels.
The kernels also have been optimized using 'Bit Hacks'. In the benchmark, the kernel (v_v_v_d) reduces its time approx. in 5us in all of its benchmarks, while the kernel (v_v_v) increments its time approx. in 7us due to the bitmask processing. 
https://github.com/BlazingDB/libgdf/blob/729cbfac6ae2281894b99644c58643184bd18d85/src/binary-operation/jit/code/kernel.cpp#L82

src/binary/binary2/kernel.cpp
ironbit: Implemented the remaining scalar-vector operations.

PR #99, opened 2018-08-15
[REVIEW] User defined selection between hash & sort based inner/left join

### LINKED ISSUE #74: Shouldn't need "-DHashJoin" as a cmake option
https://github.com/gpuopenanalytics/libgdf/blob/master/CMakeLists.txt#L94

Hash joins should always be enabled by default, and sort/hash based joins should both be exposed.

kaatish:
- Added gdf_context pointer as a parameter for inner/left single column join.
- The flag_method property of the struct can be set to either GDF_HASH or GDF_SORT which would decide the underlying join implementation to be run. GDF_SORT method runs only when flag_sorted property of gdf_context is set to true otherwise an error is returned.
- Pytests and Google test for the changes are in place.

Fixes #74


CMakeLists.txt
```diff
@@ -111,8 +111,10 @@ cuda_add_library(gdf SHARED
     src/binaryops.cu
     src/bitmaskops.cu
     src/column.cpp
+    src/context.cpp
     src/cudautils.cu
     src/datetimeops.cu
+    src/context.cpp
```
scopatz: The same file should not be added twice, because it will then be compiled and added twice.

scopatz:
Looks fine to mee, but there are a lot of API changes here.  Do we know that these won't affect pygdf?

CMakeLists.txt
kaatish: Fixed.

dantegd:
@scopatz The necessary changes to pygdf are being worked on, I already have a good portion of the work done and will be pushed soon to pygdf to avoid conflicts with the new APIs.

scopatz:
Ahh OK, great!

kkraus14:
Let's hold off on merging this PR until we have the relevant PyGDF PR ready to be merged as well. That way we can minimize the issues in having incompatible builds between libgdf and pygdf.

kkraus14:
@nsakharnykh @mt-jones we have the bindings PR ready to go here: https://github.com/gpuopenanalytics/pygdf/pull/203 so feel free to merge this when you're ready.

PR #100, opened 2018-08-16
[REVIEW] Inner & Left Join API cleanup

kaatish:
- Single column inner and left joins have been replaced with multi column data type agnostic function calls i.e. gdf_(inner/left)_join_(generic/i8/i16..) are replaced with gdf_(inner/left)_join.
- Google Test and Python Tests have been changed to reflect the new API.


kkraus14:
@nsakharnykh @mt-jones we have the bindings PR ready to go here: gpuopenanalytics/pygdf#203 so feel free to merge this when you're ready.

PR #101, opened 2018-08-16
[REVIEW] Fixes #81: remove Python 2.7 builds, add CUDA 9.2 builds, default to CUDA 9.2

### LINKED ISSUE #81: Drop Python 2.7 support
Seeing as numerous other major projects (some of which we depend on) have plans for dropping Python 2.7 support, we should drop Python 2.7 support now while we're still under heavy development and haven't started doing formal releases yet.

If I receive positive feedback (or no feedback) to this issue I'll follow up with a WIP pull request removing the Python 2.7 builds from CI.

kkraus14:


kkraus14:
It looks like Python 3.7 builds aren't really available via conda / pip yet so I'm going to remove Python 3.7 and stick to just 3.5 and 3.6.

kkraus14:
@mike-wendt Can you review before merge?

PR #104, opened 2018-08-21
[REVIEW] Remove pytest-ci option from CMakeLists.txt

mike-wendt:
`pytest-ci` option was intended to produce a `junit.xml` file for gpuCI to consume; however, it did not work. Instead of leaving this dead code in the `CMakeLists.txt` I'm removing it.

For gpuCI I am calling `py.test` directly with the flags needed to generate the file in an accessible place to report the results.


scopatz:
LGTM!

PR #106, opened 2018-08-22
[REVIEW] Replace function

gcca:
API proposal for the replace function.


GPUtester:

mike-wendt:
ok to test

scopatz:
API seems reasonable to me

src/tests/replace/replace-benchmark.cu
```diff
@@ -0,0 +1,110 @@
+/*
```
nsakharnykh: This should be moved to a separate `bench` folder as we discussed in the binary ops PR review?

nsakharnykh: And separated into benchmarks and unit tests?

src/tests/replace/replace-test.cu
```diff
@@ -0,0 +1,143 @@
+/*
```
nsakharnykh: Should add more documentation/comments to the tests

nsakharnykh (changes requested):

src/tests/replace/replace-benchmark.cu
williamBlazing: done

src/tests/replace/replace-test.cu
williamBlazing: done

src/replace.cu
```diff
@@ -0,0 +1,174 @@
...
+            const gdf_column *to_replace,
+            const gdf_column *values) {
+    if (NotEqualReplacementSize(to_replace, values)) {
+        return GDF_COLUMN_SIZE_MISMATCH;
+    }
+
+    if (NotSameDType(column, to_replace, values)) { return GDF_CUDA_ERROR; }
+
+    switch (column->dtype) {
+#define WHEN(DTYPE)                                                            \
```
harrism: I don't really like calling this macro "WHEN" -- it's not descriptive. Something like `REPLACE_CASE` would be clearer. Then you would have 

```
switch (column->type) {
    REPLACE_CASE(INT8);
    REPLACE_CASE(INT16);
    // etc.
}
```

include/gdf/cffi/functions.h
```diff
@@ -740,3 +740,15 @@ gdf_error gdf_quantile_aprrox(	gdf_column*  col_in,       //input column;
...
+
+/* replace */
+
+/// \brief Replace `to_replace` data of `column` with `values`
+/// \param[in/out] column data
+/// \param[in] to_replace contains values of column that will be replaced
+/// \param[in] values contains the replacement values
+///
+/// Note that `to_replace` and `values` are related by the index
+gdf_error gdf_replace(gdf_column *      column,
```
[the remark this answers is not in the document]
harrism: Now I see. The meaning of "replace" is not clear. It's not a copy. The semantics are actually those of "find_and_replace_all": For each value in `to_replace`, find all instances of that value in `column` and replace it with the corresponding value in `values`.  This should be made clear in the header documentation for the function. Consider changing the name to gdf_find_and_replace_all() or something like that...

harrism (changes requested):
Consider changing the function name or at least improving the documentation of what it does. One other minor suggestion.

src/tests/replace/replace-test.cu
harrism: @nsakharnykh Can you approve your review since this is resolved?

nsakharnykh (approved):

src/bench/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+
```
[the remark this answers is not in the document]
gcca: Since no other PR depends of this, I put google benchmark here. But I could make another PR if it's relevant.

PR #112, opened 2018-08-24
[REVIEW] Robust memory estimation is now in place

ogreen:
For A columns that are much bigger then B column, we have a new estimator in place that is estimating the amount of joined elements based on the results of join of B with a subset of A (that is initially the size of B).


src/join/hash/join_compute_api.h
```diff
@@ -125,105 +124,164 @@ cudaError_t compute_hash_join(mgpu::context_t & compute_ctx,
...
+  // When the ratio is big enough, we will take a subset of A equal in length to B and probe (without writing outputs). We will then approximate
+  // the number of joined elements as the number of found elements times the ratio.
+  size_type leftSize = left_table.get_column_length();
+  size_type rightSize = right_table.get_column_length();
+
+  size_type leftSampleSize=leftSize;
+  size_type size_ratio = 1;
+  if (leftSize > 5*rightSize){
+  	leftSampleSize=rightSize;
+  	size_ratio = leftSize/rightSize + 1;
```
scopatz: inconsistent spacing around `=`

scopatz:
Thanks for putting this in @ogreen!  Stylistically, there are a few oddities an inconsistencies, but the code itself seems fine otherwise.

Seems that CI is failing to install boost on one platform, though:

```
$ sudo apt-get install libboost-all-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package libboost-all-dev
```

PR #125, opened 2018-09-05
[WIP] Install gcc 7.2 using conda for TravisCI [skip-ci]

kkraus14:
EDIT: Updating to gcc 7.2 since conda and conda-forge are standardizing around this ABI version.


scopatz:
LGTM! Thanks

kkraus14:
In looking at this, I'm not sure this is the best idea... If we use GCC 5.4 it will break ABI compatibility with other conda packages.

scopatz:
I believe that `defaults` has migrated to the new compilers and ABI.  conda-forge is in process with this migration right now.

kkraus14:
@scopatz For ABI compatibility with `defaults` and `conda-forge` which version(s) of GCC should we be planning to use in the future?

scopatz:
@kkraus14 - `defaults` is a multichannel containing `main`, `free`, `pro` angd `r`. Priority is given to `main`.

* The `main` channel is built with the new ABI
* The `free` channel is built with the old ABI
* `pro` is deprecated
* `r` is a mix, depending on the R version, which we don't care about here.

Conda-forge is in the process of migrating.  Currently the `main` label uses the old ABI, while the `gcc7` label uses the new ABI. To install from conda-forge using the new ABI, you'd use:

```
$ conda install -c conda-forge/label/gcc7 <pkg>
```

Eventually, everything on the gcc7 label will be moved to the main label, but you can use the above for the time being, if the package is available.

mike-wendt:
Dug into this with my own docker container and was able to get builds working past where they are now without the CMake mods proposed on this PR. TLDR use PPAs to install gcc-7/g++-7, then setting CC/CXX env var with the path, install cmake with conda, activate root conda env, run builds

Found that CUDA 9.0/9.1 are limited to gcc6 due to NVCC. CUDA 9.2 does allow for gcc7 compiling, but I get warnings treated like errors like: https://gist.github.com/mike-wendt/4e24ac86adcc7d518518139a4934c77b

Here are the PPAs needed on 14.04:
```
apt-get update
apt-get install software-properties-common # if add-apt-repository doesn't exist
add-apt-repository ppa:jonathonf/gcc-7.2 -y
add-apt-repository ppa:jonathonf/gcc -y
add-apt-repository ppa:jonathonf/binutils -y
add-apt-repository ppa:jonathonf/debhelper -y
add-apt-repository ppa:jonathonf/nvptx-tools -y
apt-get update
apt-get install gcc-7 g++-7
```

mike-wendt:
Closing as this PR is broken and Travis will not build from it any more will continue work in #171 or whatever the successful PR to fix this issue is

PR #151, opened 2018-09-26
[REVIEW] Fix gmock failure

sklam:
Pin to googletest 1.8.1


scopatz:
Seems reasonable to me! Thanks @sklam!

PR #163, opened 2018-10-05
[REVIEW] Disable git clone depth so that GIT_DESCRIBE_NUMBER is correct for bu…

sklam:
…ild number

I noticed that build number has decreased (see https://anaconda.org/gpuopenanalytics/libgdf/files).
This should fix it by removing the clone depth by TravisCI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ dist: trusty
 sudo: required
 language: cpp
 
+git:
+  depth: false
+
 branches:
     only:
         - master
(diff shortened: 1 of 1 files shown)


scopatz:
LGTM!

sklam:
I want to wait for TravisCI to be done to be sure it works.

scopatz:
yep, of course

kkraus14:
Thanks! Merging.

PR #171, opened 2018-10-18
[WIP] Update Travis CI to gcc5

mike-wendt:
Related to #125 but unable to test with Travis CI anymore due to lack of building on push.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,14 +16,24 @@ env:
 
 matrix:
   include:
-    - env: CUDA=9.0.176_384.81
-    - env: CUDA=9.1.85_387.26
-    - env: CUDA=9.2.148_396.37
-    - env: CUDA=9.0.176_384.81 BUILD_CFFI=1 PYTHON=3.6
-    - env: CUDA=9.0.176_384.81 BUILD_CFFI=1 PYTHON=3.5
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-5
+      env: 
+        - CUDA=9.0.176_384.81
+        - CUDA=9.1.85_387.26
+        - CUDA=9.2.148_396.37
+        - CUDA=9.2.148_396.37 BUILD_CFFI=1 PYTHON=3.6
+        - CUDA=9.2.148_396.37 BUILD_CFFI=1 PYTHON=3.5
 
 
 before_install:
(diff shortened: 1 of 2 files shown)


mike-wendt:
https://docs.travis-ci.com/user/languages/cpp/#gcc-on-linux < docs from Travis on how to set the gcc version. This method gives gcc/g++ 5.5

mike-wendt:
Have build errors that I need help diagnosing as to if they are a result of gcc5.5 or something else https://travis-ci.org/rapidsai/libgdf/builds/443033699

mike-wendt:
So gcc7 builds for 9.2 but fails with an unused functions with warnings as errors https://travis-ci.org/rapidsai/libgdf/jobs/443047959

Thoughts on why this is happening?

mike-wendt:
All passed and ready for review. @kkraus14 @nsakharnykh @sklam can we review this ASAP?

kkraus14:
@mike-wendt Should we go ahead and drop support for CUDA 9.0 and 9.1 with these changes? Given the builds we're going to publish will be CUDA 9.2+ anyway I don't see the value in building for CUDA 9.0 / 9.1.

mike-wendt:
@kkraus14 Definitely could do that, I'm happy to do it either way. I guess the underlying question is for conda builds should it just be CUDA 9.2 or more than 9.2 pushed? CUDA 9.0 does give us the most compatibility at the moment. Could always revert to it for now with libgdf and then for the cuDF merger we switch to CUDA 9.2+ officially.

kkraus14:
Given we've made the decision to only support CUDA 9.2+ so that we can move to a newer GCC version I think we should just push CUDA 9.2 now. The conda team is working on a way to detect the driver version / cuda toolkit version and allow you to pull versions based on that but it isn't ready yet as far as I know. @sklam Any input you'd want to add here?

.travis.yml
```diff
@@ -19,29 +19,44 @@ matrix:
     - env: CUDA=9.0.176_384.81
     - env: CUDA=9.1.85_387.26
```
kkraus14:
```suggestion

```

.travis.yml
```diff
@@ -19,29 +19,44 @@ matrix:
     - env: CUDA=9.0.176_384.81
```
kkraus14:
```suggestion

```

kkraus14:
```suggestion

```

PR #175, opened 2018-10-22
[REVIEW] Added additional parameters to gdf_context_view for quantile bindings

dantegd:
Added function `gdf_context_view_augmented` needed for initializing parameters of `gdf_context` used in `quantile` functions.

### DIFF
--- a/include/gdf/cffi/functions.h
+++ b/include/gdf/cffi/functions.h
@@ -83,6 +83,9 @@ gdf_error gdf_column_free(gdf_column *column);
 gdf_error gdf_context_view(gdf_context *context, int flag_sorted, gdf_method flag_method,
                            int flag_distinct);
 
+gdf_error gdf_context_view_augmented(gdf_context *context, int flag_sorted, gdf_method flag_method,
+                           int flag_distinct, int flag_sort_result, int flag_sort_inplace);
+
 /* error handling */
 
 const char * gdf_error_get_name(gdf_error errcode);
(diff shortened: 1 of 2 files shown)


include/gdf/cffi/functions.h
```diff
@@ -83,6 +83,9 @@ gdf_error gdf_column_free(gdf_column *column);
 gdf_error gdf_context_view(gdf_context *context, int flag_sorted, gdf_method flag_method,
                            int flag_distinct);
 
+gdf_error gdf_context_view_augmented(gdf_context *context, int flag_sorted, gdf_method flag_method,
```
harrism: What does "augmented" refer to here? It looks like the list of options for the view is augmented, not the view itself?  If this is just an extended version of the `gdf_context_view` API, might consider redesigning that API rather than adding a new one...

Would it be invasive / onerous to just replace `gdf_context_view` with this new version?

dantegd: Was just following the existing `gdf_column_view_augmented1`, but to be honest It wouldn't be too invasive to do the replacement, will do the change early tomorrow (so late today for you I guess) and update this PR.

harrism: Perhaps `gdf_column_view_augmented` needs renamed too? In my experience usually extended APIs use "ext" on the end, not "augmented", and are usually added when you can't change the base API directly. (Is that the case here?)  "Augment" is more "supplement" or "increase" than it is "extend" -- at least in my connotation.

dantegd: was planning of eliminating `gdf_context_view_augmented` and just add the additional parameters to `gdf_context_view`, and update the code that calls that to pass all the parameters, so I don't think augmented or ext is necessary here.

harrism: Yeah, I was more referring to `gdf_column_view[_augmented]`.  Outside the scope of this PR though.

harrism:
I assume this affects the pygdf side. Is there a corresponding PR?

harrism:
rerun tests

PR #183, opened 2018-10-25
[REVIEW] Approved refactor of libgdf for merge into cuDF

mike-wendt:
Target is to merge libgdf into cudf as a subfolder, and maintain all history. This PR is a method to prepare and test the refactoring before the merge occurs.

Also this gives a path forward for outstanding PRs with the file renames and changes included here.


mike-wendt:
Same content as #180 and approved by @mt-jones 

Will correctly merge after testing completes

harrism (changes requested):
Two files are duplicated: sqls_rtti_comp.hpp and sql_set_ops.hpp should be in libgdf/src, not in libgdf/include. Delete the two files from libgdf/include and I think you are good to go.

mike-wendt:
@harrism removed duplicated files, that was a mistake in my merge of your RMM. Thanks for catching it!

harrism (approved):

